PR #24, opened 2024-06-17
RFC-14: MCS: Adding budget limit thresholds to endpoints for SC Donation

lsf37:
Original [Jira issue and discussion](https://sel4.atlassian.net/browse/RFC-14)

[Rendered version](https://github.com/seL4/rfcs/blob/0140-budget-limit-threshold/src/proposed/0140-budget-limit-threshold.md) of the RFC text.

### DIFF
--- a/src/proposed/0140-budget-limit-threshold.md
+++ b/src/proposed/0140-budget-limit-threshold.md
@@ -1,0 +1,469 @@
+<!--
+  SPDX-License-Identifier: CC-BY-SA-4.0
+  Copyright 2023 UNSW
+-->
+
+# MCS: Budget limit thresholds on endpoints for SC Donation
+
+- Author: Mitchell Johnston
+- Proposed: 2023-08-15
+
+## Summary
+
+This is a two stage RFC that proposes to add a new `budget threshold' to
+endpoints, which would only allow scheduling context donation to occur if the
+SC's available budget exceeds the threshold. This would provide a mechanism to
+prevent budget-expiry from occurring in passive servers.
+
+As a second optional stage, this threshold value could also restrict the budget
+that a passive server is permitted to consume on a donated SC. We refer to this
+as a `budget limit'. This would help bound priority inversion in a system as
+well as reducing the trust that clients must place in passive servers they call.
+
+## Motivation
+
+### Thresholds
+
+This change regarding thresholds is motivated by a desire to prevent budget
+expiry from occurring inside passive servers, improving worst-case performance
+and response time analysis. Additionally, it would shift timeout exceptions into
+being a true error case, rather than an unavoidable, but routine occurrence on
+many systems. This is likely to simplify system design.
+
+Currently, there are almost no restrictions on the amount of budget required to
+enter a passive server. This almost inevitably leads to budget expiry, resulting
+in undesirable longer blocking times for other clients of the passive server.
+This must also be accounted for in schedulability analysis of a system and is
+particularly important for servers with clients of varying criticality.
+
+The kernel currently supports a mechanism to manage budget expiry, timeout
+handlers. These allow an error-handling thread to intervene and resolve the
+budget expiry. However, as they address the issue after it occurs, these have
+some limitations. Timeout handlers can provide extra budget, however, it is not
+preferable to reward clients with additional budget just because they are
+blocking the rest of the system. Alternatively, the server can be reverted to a
+previous state, however this wastes any work the server performed and can be
+expensive or impractical, depending on the nature of the passive server.
+
+This RFC proposes a new mechanism, endpoint thresholds, which aim to prevent
+budget expiry from occurring in passive servers, rather than reacting to it
+after it occurs.
+
+### Budget Limits
+
+The second stage of budget limits is motivated by the current implicit
+requirement that clients must completely trust passive servers that they call.
+Under the current model, a server can consume as much budget as it desires on a
+donated SC and is not forced to return it. A malfunctioning or malicious passive
+server can therefore run indefinitely while also possessing a higher priority
+than its clients, significantly impeding their ability to make progress.
+
+The impact of this extends to any thread with a priority between that of the
+passive server and its clients, even threads that are not themselves clients of
+the passive server. Enforcing stronger bounds on passive server SC consumption
+would reduce the level of trust threads in a system need to have in one another.
+
+## Guide-level explanation
+
+### Thresholds Guide
+
+This change involves adding a new attribute to endpoint objects, known as a
+threshold. By default, this threshold value would be set to zero and will have
+no effect. Therefore, existing code will not need to be updated to accommodate
+thresholds.
+
+If the threshold is set to a non-zero value, sending invocations performed on
+the endpoint will only be permitted if they donate a scheduling context with
+available budget greater than or equal to the threshold. Available budget is the
+amount of budget ready for use in the SC (the sum of refills with a release time
+greater than the current time). By setting the threshold to the WCET of the
+passive server associated with the endpoint, this guarantees that the server
+will always receive an SC with sufficient budget to complete its execution,
+preventing budget expiry from occurring. This will also ensure that timeout
+exceptions will only occur as a result of a true error, such as a misconfigured
+threshold value or a fault in the server code.
+
+If a client has insufficient available budget in its SC, its refills will be
+deferred and merged until its head refill has sufficient budget at some point in
+the future. Once the thread resumes, the IPC invocation will transparently
+continue. If a client's maximum budget is less than the threshold value, an
+error will instead be returned to the client.
+
+The time required to check an SC's available budget and to defer and merge are
+both bounded by the maximum number of refills in an SC. When designing a system
+using an endpoint with a threshold, this will be an additional factor to
+consider regarding the optimal choice for the maximum number of SC refills.
+
+The threshold value of an endpoint can be set via an object invocation,
+performed indirectly through an invocation on the CNode, similar to the existing
+Cancel Badged Sends invocation. However, we restrict the right to set the
+endpoint threshold to only the original unbadged capability.
+
+Changes to the threshold value are propagated weakly, meaning that clients only
+interact with the threshold value at the moment where they invoke the endpoint.
+If a threshold value is increased, clients already enqueued on the endpoint are
+unaffected, even if they now possess insufficient budget. Similarly, threads
+which have had their budget deferred will not be woken up early, even if the
+threshold value is reduced. As the threshold is intended to reflect the WCET of
+the passive server, changing it would be a rare operation. Therefore, we
+consider this weak propagation behaviour acceptable.
+
+Additionally, invocations which do not support scheduling context donation are
+no longer permitted to invoke an endpoint with a non-zero threshold set. They
+instead fail immediately as an invalid operation. This prevents a passive server
+from performing a rendezvous with a client that did not donate an SC, which
+would result in the server being unable to run, which causes similar issues to
+budget expiry. Invoking a passive server without donating an SC is fundamentally
+an invalid operation, so it is therefore desirable for the IPC to fail upon the
+initial send, rather than causing an exception only after rendezvousing with the
+server.
+
+As an optional extension, the new defer and merge behaviour of SC's can also be
+made available to threads as an explicit system call. We have provisionally
+called this YieldUntilBudget, denoting it as a variant of Yield. This would
+allow threads to specify a desired quantity of budget, after which the kernel
+will merge and defer refills until the thread's head refill exceeds the desired
+budget. We do emphasise that this additional system call is not required for the
+functionality of thresholds. However, it will essentially be implemented
+regardless as part of thresholds and exposing it to users is an easy change.
+
+### Budget Limits Guide
+
+The optional budget limit change extends the meaning of an endpoint threshold
+value to also restrict budget consumption by a passive server on a donated SC.
+An additional bit is added to endpoint objects to toggle budget limit behaviour.
+By default, budget limits are disabled and only the aforementioned threshold
+behaviour applies. If the threshold is non-zero and budget limits are enabled,
+then both threshold and budget limit behaviour will be in effect. A budget limit
+has a no effect if the threshold is zero, regardless of the value of the budget
+limit bit.
+
+If a passive server returns a donated SC having consumed less than the budget
+limit (equal to the threshold value), then the budget limit has no effect.
+However, if the server's consumption reaches the budget limit, the kernel will
+forcibly return the SC to the caller, ensuring that the server’s consumption
+does not exceed the limit. This will almost certainly leave the server in a
+stuck state, so the server's timeout handler will be invoked to recover it.
+
+When an SC is donated over an endpoint with budget limits enabled, that SC is
+marked as possessing a budget limit. An SC with a budget limit can only be
+further donated over endpoints that also have budget limits set.
+
+During such a donation, the kernel must still determine whether the donated SC
+has a sufficient budget (greater than the threshold). However, rather than
+considering the total available budget in the SC, the kernel considers the
+budget remaining until the existing budget limit would be reached. This allows
+the kernel to guarantee that every server is able to use its full budget limit
+allocation, while also being able to enforce the limit if any server exceeds its
+allocated budget limit. The SC is only marked as no longer possessing a budget
+limit once it is returned over the original budget limit endpoint.
+
+## Reference-level explanation
+
+### Thresholds Reference
+
+This RFC proposes to add a `ticks_t` value to endpoint objects as the
+`threshold'. It is also viable to associate the threshold with endpoint
+capabilities, this is discussed in the rationale and alternative section below.
+For every sending IPC invocation, fastpath or slowpath, the kernel checks the
+threshold value of the relevant object. If it is set to zero, this represents
+threshold behaviour being disabled and the IPC continues, with no further
+behaviour changes.
+
+However, if the threshold value is non-zero, then the kernel calculates the
+available budget of the thread (released refills in the SC) and compares against
+the budget required to pass the threshold. The intention is for the 'threshold'
+value to represent the required runtime of a passive server. Therefore, the
+budget to pass the threshold is the 'threshold' value plus twice the kernel
+WCET, to account for the call and reply system calls. Additionally, the kernel
+must account for time consumed by the client, but not yet charged to its SC. If
+the available budget is sufficient, again, the IPC continues without further
+behaviour changes.
+
+However, if the available budget is insufficient, alternate action needs to be
+taken. At this point, the fastpath and slowpath diverges. If the available
+budget check on the fastpath fails, the kernel switches over to the slowpath to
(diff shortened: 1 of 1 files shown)


kent-mcleod:
If the server thread was told how much budget was available on the donated SC when it is unblocked from seL4_Recv, then it could immediately return an error to the client based on some policy it can implement itself.  This in combination with the proposed mechanism that allows a well-behaving client to yield-until-budget before making the call gives more power to both client and server and doesn't require adding more complexity to endpoints. 

A single threshold value on an endpoint would have to be suitable for all calls over that endpoint whereas a server implementing a policy at user level could enforce a different threshold limit that's specific to each unique call. 

I couldn't find a user-level approach discussed in the proposal or the thesis or the comments.

lsf37:
I would also welcome some more thinking on alternatives to this approach. Doesn't mean we have to take them, but the design space is large and the RFC at least does not show exploration of much of it -- it's the prototype of one implementation that happens to work, but for me it has two main problems:

- To my understanding we are now introducing a loop into the IPC fastpath(!) that is a priory only bounded by the absolute maximum number of refills a user can create a scheduling context with. As the performance number tables show, that means there is no "IPC takes x cycles" for specific scenarios any more -- it all depends on how things were scheduled before and how many refills happen to be in the scheduling context. The user does not really have much control over that apart from setting the maximum number.

- The other aspect is verification cost: this is the IPC path -- even small modifications are very expensive there, and even the simplest version of the RFC is not a small modification. This would need significant verification funding first to proceed (it does not look to me like this can be made conditional easily -- if it can, then it could be made available in non-verification configs first).

Would it make sense the keep a "remaining overall budget" field in the SC that is updated each time an SC is charged and re-charged? (I.e. explicitly keep the sum of all refill budgets). This would avoid the loop and make Kent's alternative of a server (and even well-behaved client) checking the remaining budget quite fast and easy to implement without introducing new kernel mechanisms on endpoints.

lsf37:
Status update from 2024-07-25 TSC meeting: awaiting change of prototype implementation by Indan.

Indanz:
> If the server thread was told how much budget was available on the donated SC when it is unblocked from seL4_Recv, then it could immediately return an error to the client based on some policy it can implement itself. This in combination with the proposed mechanism that allows a well-behaving client to yield-until-budget before making the call gives more power to both client and server and doesn't require adding more complexity to endpoints.

The problem is that we don't know how much budget is available when the server gets unblocked. To know that, we need to take a timestamp on the fast path, which currently doesn't happen.

So from a performance point of view your proposal would require all IPC calls to take a timestamp and an extra yield-until-budget syscall for tasks wanting this feature.

With the proposal the cost is one extra if-check to check for budget threshold for all IPC calls, and taking the timestamp only when a budget is set.

Indanz:
> * To my understanding we are now introducing a loop into the IPC fastpath(!)

I'm getting rid of those new loops and will go to the slow path if the head budget isn't sufficient.

> The other aspect is verification cost

The loop to gather enough budget is the same loop as the one at the end of `refill_budget_check`, except that the budget threshold is used instead of `MIN_BUDGET`. `refill_unblock_check` already merges overlapping refills (I don't understand how they can happen though, seems easier to avoid than to fix afterwards to me).

But yes, if you want to proof that budget threshold feature does what it should, the verification cost will be high no matter how it's implemented.

> it does not look to me like this can be made conditional easily -- if it can, then it could be made available in non-verification configs first

I think it could be made conditional easily, but the question is whether you want that.

kent-mcleod:
> The problem is that we don't know how much budget is available when the server gets unblocked. To know that, we need to take a timestamp on the fast path, which currently doesn't happen.

There is currently a timestamp taken on every slowpath entry to the kernel because every time the kernel is entered it calculates whether the time remaining in the head refill is greater than `MIN_BUDGET`: `refill_head(sc)->rAmount - usage` in `refill_sufficient()`. The result of this calculation is also the same as the amount of time until the currently programmed timer interrupt will fire and is the amount of budget available to the server before a timeout fault would be generated.

If this check isn't also in the fastpath then it would need to be added regardless.

Indanz:
> If this check isn't also in the fastpath then it would need to be added regardless.

Why? We already paid most of the cost of a syscall, handling the fast path is always faster than doing anything else. `MIN_BUDGET` applies to generic system calls as it is bound to WCET, not fastpath ones, which are on the order of one microsecond.

At worst the called task will be interrupted within `MIN_BUDGET` by the timer interrupt to end its time-slice. You can usually do 10 or more IPC calls within `MIN_BUDGET` time.

IPC fastpath is a context switch without SC change, and because of that it can avoid doing anything timing related at the moment.

(For SMP I think we should take the timestamp before taking the kernel lock, see #844, and always taking the timestamp at entry would simplify some things, so I'm not against it in principle. But it will add extra overhead, especially on RISCV where it usually traps.)

kent-mcleod:
> Why? We already paid most of the cost of a syscall, handling the fast path is always faster than doing anything else. `MIN_BUDGET` applies to generic system calls as it is bound to WCET, not fastpath ones, which are on the order of one microsecond.

In the kernel design the fastpaths are each optimizations of a specific slowpath kernel operation. So if refill_sufficient() would return false in the slowpath and result in a thread switch, then the fastpath would have to produce the same thread switch, or else fall back to the slowpath. To facilitate this, the fastpath needs to perform the same check.

> IPC fastpath is a context switch without SC change, and because of that it can avoid doing anything timing related at the moment.

It arguably can't avoid performing the budget check. The kernel has to ensure there is enough remaining budget to get back into the kernel whenever it returns to user level.  It has to guard against a thread calling late enough in its time slice and amplifying its budget via a non-preeptable kernel operation.

I think there's a design inconsistency here where the fastpath is always expected to behave the same as a particular path through the slowpath, but the slowpath will now abort an operation if it doesn't have enough time to complete while the fastpath is hoped to be fastenough to always have enough time available to complete. 

Shouldn't the fastpath also abort an operation if it doesn't have enough time to complete?

We either want to assume that:
1. the fastpath time rounds to zero (and not needing the check as a consequence), 
2. the fastpath time doesn't round to zero and so a check is needed, and because of the design requirement that the fastpath follows the same behavior as the slowpath then the check is the same as the check in the slowpath.  In this case maybe we could define the check to be parameterized by input arguments to the kernel, and in the case where it's the fastpath condition the check checks a smaller time value?

Indanz:
Thinking about time is always tricky, hence my delayed response.

> It arguably can't avoid performing the budget check. The kernel has to ensure there is enough remaining budget to get back into the kernel whenever it returns to user level. It has to guard against a thread calling late enough in its time slice and amplifying its budget via a non-preeptable kernel operation.

The kernel does not check at syscall entry if there is a higher priority thread scheduled within WCET time. Hence the worst case expected scheduling delay is equal to WCET. Even if a task calls a fastpath syscall just before it's timeslice ends, the worst case scheduling delay for the next task is much less than WCET.

The budget amplification is negligible. However, avoiding that amplification will not reduce the total amount of kernel operation time, though it might avoid taking a timer interrupt. I'm not convinced that stealing `MIN_BUDGET` budget from a task is better that letting it overrun by a fraction.

> We either want to assume that:
> 
>     1. the fastpath time rounds to zero (and not needing the check as a consequence),

This is the assumption I am making. Scheduling is at 1 us granularity currently, so anything close to that is a rounding error.

>     2. the fastpath time doesn't round to zero and so a check is needed, and because of the design requirement that the fastpath follows the same behavior as the slowpath then the check is the same as the check in the slowpath.  In this case maybe we could define the check to be parameterized by input arguments to the kernel, and in the case where it's the fastpath condition the check checks a smaller time value?

I have the feeling that the kernel currently abuses WCET for kernel entry and exit overhead, while that should be separate values from WCET (the only explanation I have for `MIN_BUDGET = 2 * WCET`). As WCET is used everywhere, it's unclear what purpose each margin has.

In the end it's about what timing guarantees seL4 provides. In my view seL4 should be suitable for hard-realtime, which implies (ignoring IRQs):

1) Tasks start when they should. This currently has error `Es = WCET`.
2) Tasks get enough budget in a timely manner. This currently has error of at least `Eb = Es + MIN_BUDGET` because of the budget check. But because notifications cause [LIFO scheduling behaviour](https://github.com/seL4/seL4/issues/902), no guarantees whatsoever can be given at the moment.
3) Scheduling errors should not accumulate. `Es` and `Eb` are non-accumulating. (Notifications causing LIFO scheduling is exactly such a big problem because it does accumulate, but let's ignore it for this analysis.)

Note that no budget overruns happening is not a design goal in itself. It may be necessary to be able to make certain guarantees, but I'm not fully convinced it is yet. It does avoid accumulating scheduling delays in case other tasks of the same priority all do budget amplification. However, that error is at most WCET per task (much less in reality), while the budget check causes tasks to lose up to `MIN_BUDGET` per period. I'd argue that a scheduling jitter of `MIN_BUDGET` is much worse than a scheduling jitter bound by WCET and much less in practice.

In other words, currently tasks needing a budget of `b` need to set `budget = b + Eb` if they have a time limit and ` budget = b + MIN_BUDGET` otherwise. Without the budget check they can set `budget = b + Es` or `budget = b` for all tasks. Overall time can be divided more efficiently without a budget check at each syscall entry and scheduling jitter will be lower. As far as I can remember, the original sporadic scheduling paper did not check budget at every kernel entry either.

Indanz:
To clarify, I think a slight budget overrun is fine if it means the task will get that amount less budget next period, to avoid accumulating errors.

Indanz:
Simplified implementation of budget thresholds can be found here: https://github.com/Indanz/seL4/pull/2

src/proposed/0140-budget-limit-threshold.md
```diff
@@ -0,0 +1,469 @@
...
+### Thresholds
+
+This change regarding thresholds is motivated by a desire to prevent budget
+expiry from occurring inside passive servers, improving worst-case performance
+and response time analysis. Additionally, it would shift timeout exceptions into
+being a true error case, rather than an unavoidable, but routine occurrence on
+many systems. This is likely to simplify system design.
+
+Currently, there are almost no restrictions on the amount of budget required to
+enter a passive server. This almost inevitably leads to budget expiry, resulting
```
pingerino: nit: this almost inevitably leads to budget expiry
suggestion: this can lead to undesirably budget expiry in certain system designs

Not really invevitable, that's up to the system. If you have a system where overruns are tolerated, or where you think you have the WCET mostly sorted but not completely, you can have occasional budget expiry and handle it. Whether it's desirable or not is up to the system design and is a matter of policy. That said, *if* you wanted a system where you really did not want timeout exceptions,

src/proposed/0140-budget-limit-threshold.md
```diff
@@ -0,0 +1,469 @@
...
+As a second optional stage, this threshold value could also restrict the budget
+that a passive server is permitted to consume on a donated SC. We refer to this
+as a `budget limit'. This would help bound priority inversion in a system as
+well as reducing the trust that clients must place in passive servers they call.
+
+## Motivation
+
+### Thresholds
+
+This change regarding thresholds is motivated by a desire to prevent budget
```
pingerino: suggestion: this would be better worded as "to provide an option for users to configure passive servers without requiring timeout handlers"

src/proposed/0140-budget-limit-threshold.md
```diff
@@ -0,0 +1,469 @@
...
+budget expiry. However, as they address the issue after it occurs, these have
+some limitations. Timeout handlers can provide extra budget, however, it is not
+preferable to reward clients with additional budget just because they are
+blocking the rest of the system. Alternatively, the server can be reverted to a
+previous state, however this wastes any work the server performed and can be
+expensive or impractical, depending on the nature of the passive server.
+
+This RFC proposes a new mechanism, endpoint thresholds, which aim to prevent
+budget expiry from occurring in passive servers, rather than reacting to it
+after it occurs.
```
pingerino: The motivation here would be improved greatly by adding that this is additional, demostrating why it's needed, without saying all the other options are bad. Reacting is a legitiamte way to build a system, especially if your WCET is very loose. They have legitimate use cases depending on client policy.

src/proposed/0140-budget-limit-threshold.md
```diff
@@ -0,0 +1,469 @@
...
+using an endpoint with a threshold, this will be an additional factor to
+consider regarding the optimal choice for the maximum number of SC refills.
+
+The threshold value of an endpoint can be set via an object invocation,
+performed indirectly through an invocation on the CNode, similar to the existing
+Cancel Badged Sends invocation. However, we restrict the right to set the
+endpoint threshold to only the original unbadged capability.
+
+Changes to the threshold value are propagated weakly, meaning that clients only
+interact with the threshold value at the moment where they invoke the endpoint.
```
pingerino: question: There's also an implication here that all of the server's operations are bounded by the same amount. How practical is that?

src/proposed/0140-budget-limit-threshold.md
```diff
@@ -0,0 +1,469 @@
...
+called this YieldUntilBudget, denoting it as a variant of Yield. This would
+allow threads to specify a desired quantity of budget, after which the kernel
+will merge and defer refills until the thread's head refill exceeds the desired
+budget. We do emphasise that this additional system call is not required for the
+functionality of thresholds. However, it will essentially be implemented
+regardless as part of thresholds and exposing it to users is an easy change.
+
+### Budget Limits Guide
+
+The optional budget limit change extends the meaning of an endpoint threshold
```
pingerino: question: can there be some use cases here to illustrate what this is for?

src/proposed/0140-budget-limit-threshold.md
```diff
@@ -0,0 +1,469 @@
+<!--
```
pingerino: I'm coming to this PR very late, since it's being reviewed by the TSC this week.  I'm not vehemently opposed to the change but I'm not yet convinced by the motivation and would love to see my questions answered. It's great to see work happening on the MCS api and I love to see it develop and evolve.

pingerino (commented):
I've done a high level pass due to the upcoming TSC meeting. 
I don't have a lot of time to look at this.

On the surface adding thresdholds might be a useful change, though I would really like to see some use cases  to motivate the effort involved in merging and verifying the change.  

I'm happy to discuss this further. I've used conventional comments to make my comments, and tried my best to not come across as unsupportive, while also being appropriately sceptical. Great to see work being done to further the MCS apis.

Indanz:
> I'm not necessarily against changing the preemption point behaviour in MCS, but it is unclear to me what consequences it has.

It will make pre-emptible sycalls run longer because of the extra overhead.

> It seems to me like this would change behaviour slightly (exit, take IRQ immediately, re-enter kernel).

Usually there is no pending IRQ and it would just restart the current syscall.

> It also sounds to me like this would break the kernel entry/exit budget computations, because a single kernel call could now have the cost of entry -> preemption -> exit -> immediate entry -> IRQ handling/schedule new thread -> exit, whereas before it was entry -> preemption -> IRQ handling/schedule -> exit.

No, it just does the usual cost of entry -> preemption -> exit. The budget is updated and checked at the next entry as done normally for all syscalls.

It is the current code that is weird and duplicates normal bookkeeping and IRQ checking in the pre-emption code to avoid running out of budget.

> To be fair, I'm not sure what the point of the entry/exit WCET is that MCS is doing, because it does not reflect the actual WCET and any thread can go over its budget roughly by the amount of the actual kernel WCET. If we have that, why do we try so hard to account for just entry/exit?
> 
> Should we just admit that we can't fully guarantee absence of budget overrun by kernel WCET and just leave that up to user-level analysis? That would simplify a lot of things in the MCS implementation in general. (It would also make a lot of verification work obsolete that has already been put in to do verify all that complexity, but that's not a good reason to not get rid of it).

Agreed with all of that. I think the harm of enforcing MIN_BUDGET is bigger than the gain, see my discussion with Kent above. But we should continue this discussion in a new issue.

> I'm also not entirely sure about adding invocations to endpoints in general.

I'm not doing that, I'm just saying it's a complication that needs some solution.

> There probably are some points in that design space (I haven't fully checked Indan's draft for this), but it does feel like a hack. Is there nothing cleaner that we can do?

My hacky solution is making it a reply object invocation. The rationale being that for calls you need access to an EP cap and a reply cap anyway, and that this should be under the server's control.

The original RFC proposed a CNode invocation as a work-around, but that's worse I think.

Another way is using a magic nonsensical combination of flags values in `seL4_MessageInfo_t` to alter the behaviour of normal endpoint calls, but that seemed very ugly.

pingerino:
> I have the strong impression that WCET as used in MIN_BUDGET started as kernel entry/exit time, but ended up being WCET instead. Why else would MIN_BUDGET be twice WCET?

Yes, it was supposed to be enough time for a thread to get into and out of the kernel. Given we only theoretically had one estimate, this rounds up to 2 * WCET.

Indanz:
> A note on MIN_BUDGET: some practical experiments are needed here to quantify the impact on latency and precision if you don't set this. E.g, if a schedulable set of tasks is set up, what time does the highest prio task end up waking up. 

I did a simple analysis in an earlier [comment](https://github.com/seL4/rfcs/pull/24#issuecomment-2308437417) above.

My practical experience is with a system with budgets in the order of 100us and a WCET of 10us, and then it causes a lot more scheduling jitter (as measured with RTP audio quality under heavy load). Setting WCET to 1 improved system behaviour and reduced jitter. (Sadly I don't have access to that system any more, so can't repeat the tests easily.)

Indanz:
A big question is whether the budget limit should also apply to round-robin tasks or only to sporadic tasks:

- On the one hand, RR tasks have no control over when they run out of the current budget and doing this would introduce random scheduling jitter and overhead. This is the reason why timeout faults don't happen for RR tasks. Not applying budget limit for RR tasks would be consistent with this. Not enforcing the limit would also limit the overhead to one extra if-check, making only sporadic tasks pay the price.
- On the other hand, it would (usually) avoid budget amplification compared to other equal priority tasks not doing calls to higher priority servers. (See https://github.com/seL4/seL4/issues/1365). It also opens the possibility to enable timeout faults for RR task calls to passive servers.

That said, Budget Thresholds are not very user friendly to use. It's great if you know actual timings and deadlines, but it's hard to use if you don't. The user needing to do this kind of extra work to avoid unintended starvation or budget amplification feels wrong though.

An alternative solution for sporadic tasks causing problems by doing passive calls is to simply prohibit such calls.

Then the user either needs to make servers used by sporadic tasks active, or modify the SC before and after doing such calls, or add an intermediate active server which forwards the call. Another option is to modify seL4 so either users can bind a special SC for such calls, or to automatically convert the sporadic task to round-robin during passive calls. (Idea: Move the Notification SC to the TCB and use it for both handling notifications in passive servers and for sporadic tasks doing IPC calls to passive servers.)

An alternative solution for RR amplifying budget is mentioned in the above issue already. That solution would limit the amplification to a once-off event during the passive call, and penalise the over budget task afterwards to make up for it. We should still do this even with Budget Thresholds, because if the configured budget is too short, the original issue will resurface. Once this is done, Budget Thresholds don't add much additional benefit any more.

Another big problem of Budget Thresholds is that it only works if all clients have more budget than the worst case execution time of the passive server.

PR #26, opened 2024-06-21
RFC-18: FPU switching

lsf37:
We propose to let user level configure policy on when FPU context is switched instead of relying on runtime faults to switch FPU context lazily. Policy is configured via a flag in the TCB, authorised by the corresponding TCB capability.

This improves performance, moves policy from kernel to user, enables verification of FPU handling, and fixes an information leakage in current FPU context switching.

[Rendered version](https://github.com/seL4/rfcs/blob/f5500fe/src/active/0180-fpu-switching.md).

### DIFF
--- a/src/proposed/0180-fpu-switching.md
+++ b/src/proposed/0180-fpu-switching.md
@@ -1,0 +1,378 @@
+<!--
+  SPDX-License-Identifier: CC-BY-SA-4.0
+  Copyright 2024 Proofcraft Pty Ltd
+-->
+
+# FPU Context Switching
+
+- Author: Gerwin Klein, Rafal Kolanski, Indan Zupancic
+- Proposed: 2024-06-21
+
+## Summary
+
+We propose to let user level configure policy on when FPU context is switched
+instead of relying on runtime faults to switch FPU context lazily. Policy is
+configured via a flag in the TCB, authorised by the corresponding TCB
+capability.
+
+This improves performance, moves policy from kernel to user, enables verification
+of FPU handling, and fixes an information leakage in current FPU context
+switching.
+
+## Motivation
+
+Current FPU context switching in seL4 disables the FPU on kernel exit, waits for
+a thread to fault when it accesses the FPU, transparently handles the fault,
+saves/restores FPU context, and then returns to the thread.
+
+The motivation for this scheme was that threads that never use the FPU do not pay
+for FPU context switching cost. However, it has the following problems:
+
+- this is a fixed context switching policy in the kernel;
+
+- most of this complexity is invisible to the current formal verification, and
+  therefore not covered by the proofs;
+
+- when switching between domains, this scheme introduces information flow
+  leakage, because the kernel performs actions on behalf of domain A while
+  running in domain B: assuming A and B are the only domains, the first thread
+  running in domain B can determine whether a thread in domain A used the FPU
+  simply by using the FPU itself, since kernel FPU save/restore actions will
+  only occur if another thread used the FPU.
+
+- performance of this scheme is not optimal -- taking a fault every time the FPU
+  is used adds unnecessary overhead, especially in systems where FPU usage is
+  high.
+
+The proposed scheme fixes these problems and maintains the original intent that
+threads which do not use the FPU should not pay FPU context switching cost.
+
+The assumption is that application threads are likely to be using the FPU, and
+that low-level OS threads are unlikely to be using the FPU. One particular case
+we are aiming to cover is an application thread calling the OS and the OS
+returning to that thread. Ideally, this case should not need to save or restore
+any FPU registers.
+
+## Guide-level explanation
+
+### Concepts
+
+- The TCB will get a new flag `fpuDisabled` that determines whether the FPU is
+  disabled or enabled when the kernel switches to the thread. If the FPU is
+  disabled and the thread  uses the FPU, a fault is generated and delivered to
+  the thread's fault handler like other faults. If the FPU is enabled, FPU
+  context is saved and restored as needed. There is no more kernel-level
+  transparent FPU fault handling.
+
+- For backwards compatibility, the flag is initially false (the FPU is enabled)
+  on platforms that have FPU enabled at kernel configuration time.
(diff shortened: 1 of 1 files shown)


src/proposed/0180-fpu-switching.md
```diff
@@ -0,0 +1,378 @@
...
+
+- most of this complexity is invisible to the current formal verification, and
+  therefore not covered by the proofs;
+
+- when switching between domains, this scheme introduces information flow
+  leakage, because the kernel performs actions on behalf of domain A while
+  running in domain B: assuming A and B are the only domains, the first thread
+  running in domain B can determine whether a thread in domain A used the FPU
+  simply by using the FPU itself, since kernel FPU save/restore actions will
+  only occur if another thread used the FPU.
```
kent-mcleod, line 41: Would there still be leakage with the current scheme on domain switches if the FPU was always reset back to an empty state on every domain switch?

Xaphiosis: No, if we prove an invariant that the current FPU owner is a thread in the current domain then that will be fine, provided we pull the rest of the FPU switching into the verified subset... but the way it's being done now is not very verification friendly.

src/proposed/0180-fpu-switching.md
```diff
@@ -0,0 +1,378 @@
...
+  out to be a common case.
+
+The proposed scheme is slightly more complex than the semi-lazy scheme with
+early context save, and early saving would not need any special treatment on
+domain switch, but the targeted use case of OS threads without FPU called by an
+application thread with FPU would still need one FPU context save and restore,
+whereas the proposed scheme does not require any.
+
+
+## Prior art
```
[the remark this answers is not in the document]
Xaphiosis: For the domain switching, the Arm-hyp version has the same confidentiality (infoflow) issue, *but* relevant aspects of VCPU switching are in the verified subset. Switching to NULL (None) VCPU on domain switch is a tweak we want to introduce in the future, but one that won't require an RFC. It would also require an invariant that the current owner of the VCPU is in the current domain.

So, this RFC would make the FPU switching like VCPU switching is currently, plus plugging the hole you pointed out, so we can re-use a bunch of the proof techniques. Does this help at all?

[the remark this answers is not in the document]
kent-mcleod: Thanks @Xaphiosis , your explanations are helpful.

src/proposed/0180-fpu-switching.md
kent-mcleod: So should we already want to make a change like what's described in (https://github.com/seL4/seL4/pull/1271) that resets the FPU state on a domain switch?

Xaphiosis: Roughly speaking yes, except that `switchFpuOwner` is not in the verified subset due to it only being previously present on the FPU fault path and paved over during X64 verification, so we'd have to have a magical machine op for it.

Aside: given that the proof abstracts over `numDomains`, it would be preferable that you check with `numDomains > 1` in an `if` and not preprocessor (see comment in `domainConstants` enum).

lsf37: Agreed, that change would break verification in an unpleasant way right now. One of the motivations for doing it properly via this RFC.

src/proposed/0180-fpu-switching.md
```diff
@@ -0,0 +1,378 @@
...
+- when switching between domains, this scheme introduces information flow
+  leakage, because the kernel performs actions on behalf of domain A while
+  running in domain B: assuming A and B are the only domains, the first thread
+  running in domain B can determine whether a thread in domain A used the FPU
+  simply by using the FPU itself, since kernel FPU save/restore actions will
+  only occur if another thread used the FPU.
+
+- performance of this scheme is not optimal -- taking a fault every time the FPU
+  is used adds unnecessary overhead, especially in systems where FPU usage is
+  high.
```
kent-mcleod, line 45: This and the line at the start of this section ("Current FPU context switching in seL4 disables the FPU on kernel exit,") could be misleading. The kernel doesn't disable the FPU on every kernel exit.  Just when the current thread is not the current owner of the FPU. If the current thread is also the current owner of the FPU, then the FPU is enabled.

src/proposed/0180-fpu-switching.md
kent-mcleod: Can you explain why it's currently considered unverifiable? The proposed scheme also requires the FPU context to be reset on a domain switch and requires tracking the current FPU owner. The behavior for handling this seems independent from how the FPU owner is set, (ie via explicit flag in TCB + context switch, or via use and fpu fault)?

src/proposed/0180-fpu-switching.md
Indanz: Agreed it should be made clearer that faulting at every FPU usage only happens when multiple threads use the FPU. On x86 this seems to be happening now because compilers are eager to use SIMD instructions even for integer operations, so in practise you get one FPU fault every context switch.

src/proposed/0180-fpu-switching.md
lsf37: It's not entirely unverifiable, just ugly and work intensive. The function `switchFpuOwner` is currently underneath the the machine interface limit -- mentioning it means we'd have to change the machine interface so that it becomes visible, which then draws in a lot of the rest of the FPU handling, which then makes it apparent in verification that there is currently a kernel entry and exit that is not treated. It would be possible, but it'd be a whole lot of work that we'd then need to throw away when we change away from fault-based lazy FPU switching. If we decide that we do want to keep the fault-based lazy switching, that's probably what we'd need to do.

src/proposed/0180-fpu-switching.md
lsf37: Have tweaked the text a bit to clarify that.

lsf37:
This RFC was accepted without at the TSC meeting on 2024-07-25.

Xaphiosis (approved):
Accepted at TSC meeting 2024-07-25

PR #30, opened 2025-03-19
Add RFC for deprecating and removing old boards

wom-bat:
We currently support and mark as fully-supported a number of boards that have been discontinued by their manufacturers.  This RFC is to remove the oldest of these from the 'Supported' list, and to start a process for removing them from the seL4 kernel.


lsf37:
The TSC approved this RFC in the meeting on 3 Dec 2026, with the exception of the Beagleboard (`omap3`), which is still available and under CI.

We decided to check for board availability/deprecation annually, but still follow the comment and RFC process to remove boards so people have a chance to reply.

lsf37 (approved):

PR #33, opened 2025-09-10
RFC-20: Runtime Domain Schedules

lsf37:
This RFC proposes to add the ability to set, maintain, and switch between domain schedules at runtime, authorised by the existing DomainCap, retaining the current information flow proof and behaviour when the DomainCap no longer exists in the system or is not used.

[Rendered version](https://github.com/seL4/rfcs/blob/0200-domain-schedules/src/proposed/0200-domain-schedules.md)


Indanz:
Initial draft implementation can be found at https://github.com/seL4/seL4/pull/1511. But please keep design discussions here. That implementation differs in details from the RFC because I didn't had the draft RFC in front of me while writing the code.

- The implementation only checks the domain length for end markers and there is no need to force zero domain. As domain 0 is also a valid domain, requiring both length and domain to be set to zero for end markers does not make much sense. If we do set requirements for the end marker domain value, an invalid domain would make more sense. But that would be less user friendly as it depends on `numDomains` and is harder to explain (and again, the actual code will just ignore the domain value).

- Another difference is that the domain time is not checked for `MIN_BUDGET`. The rationale being that if the user messes up their scheduling so much that nothing can run, then having a domain time less than `MIN_BUDGET` is the least of their problems. The other reason to not enforce it is if the tasks in such a domain make no system calls, then there is no reason to enforce `length > MIN_BUDGET`. Most `MIN_BUDGET` checks in the kernel seem arbitrary, I rather not add more of them.

- One issue I ran into was that having the time in `word_t` makes it either 32-bit or 64-bit. Perhaps we should make it fixed size, either always 32-bits or 64-bits instead of word_t. Limiting `dom_t` to 32-bits if the length is 32-bits would also make sense. With a 1GHz timer clock, 32-bit is enough for a 4 second long domain time, so seems sufficient. Users can always concatenate entries if 32-bits isn't enough time for one domain for whatever reason.

- Now that we're breaking the API anyway, it's an excellent moment to rename `DomainSetSet` to `DomainSet`. I understand the humour, but as the function is actually changing the domain of a TCB and does not modify the set of domains, calling it `DomainSetSet` is somewhat misleading.

lsf37:
If we are standardising on ticks, making duration always `uint64_t` would definitely make sense.

Indanz:
> If we are standardising on ticks, making duration always `uint64_t` would definitely make sense.

If it was up to me, I would combine domain and duration in 64 bits, e.g. 16 bits domain and 48 bits for ticks.

Anyway, a downside of making it more than 32-bits is that retrieving the argument becomes harder, as it needs two registers on 32-bit platforms and only one on 64-bit. So no matter what we do, it will result in some ugliness somewhere. (Need something similar to `mode_parseTimeArg` on non-MCS.)

lsf37:
My summary of the discussion so far would be that we have not fully converged yet on the minor encoding details above, but nobody has voiced any objections to the overall design.

corlewis (approved):

PR #34, opened 2025-10-21
RFC-21: Unify behaviour of the Page Map invocation

Kswin01:
This RFC proposes to enforce the same over-map behaviour of the page map invocation across all architectures. That is, 
prevent over-mapping on all architectures.

A draft implementation for preventing overmapping can be found here: https://github.com/seL4/seL4/pull/1526.
The associated `sel4test` changes/additions can be found here: https://github.com/seL4/sel4test/pull/148.

### DIFF
--- a/src/proposed/0210-over-mapping.md
+++ b/src/proposed/0210-over-mapping.md
@@ -1,0 +1,91 @@
+<!--
+  SPDX-License-Identifier: CC-BY-SA-4.0
+  Copyright 2025, UNSW
+-->
+
+# Unify behaviour of the Page Map invocation
+
+- Author: Krishnan Winter
+- Proposed: 2025-10-21
+
+## Summary
+
+Enforce the same over-map behaviour of the page map invocation across all architectures.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


lsf37:
My main question on this is: why not change the behaviour of RISC-V and allow overmapping everywhere? It is the smaller change and is more permissive. And it is perfectly safe from the kernel's perspective.

Overmapping will create FrameCaps with stale mapping information, but deleting page tables already does the same, so the kernel still has to deal with the issue.

I don't think we should restrict behaviour if we don't need to.

Kswin01:
Yep, that's a valid approach as well that I mentioned in the alternatives section. I have no strong feelings either way, whether that be allowing or preventing over-mapping on all architectures. The change will be a lot simpler for allowing over-mapping on RISCV, and in my mind it would impact less existing users.

lsf37:
The TSC meeting on 3 Dec 2026 deferred approval on this RFC, but encourages progress for

- adding over-mapping checks on the remaining platforms
- while allowing access rights upgrade/downgrade
- approval will need a verification plan

When a corresponding implementation PR is available, Gerwin volunteered to estimate verification effort.

PR #35, opened 2025-10-24
RFC-22: SBI capability

Ivan-Velickovic:
This RFC proposes a new SBI capability for the purpose of accessing hardware protected by machine mode (M-Mode).

An initial implementation is available at https://github.com/seL4/seL4/pull/1532.

### DIFF
--- a/src/proposed/0220-sbi-cap.md
+++ b/src/proposed/0220-sbi-cap.md
@@ -1,0 +1,147 @@
+<!--
+  SPDX-License-Identifier: CC-BY-SA-4.0
+  Copyright 2025 UNSW
+  Copyright 2025 Proofcraft
+-->
+
+# RISC-V SBI capability
+
+- Author: Gerwin Klein & Ivan Velickovic
+- Proposed: 2025-10-23
+
+## Summary
+
+This RFC proposes to add a new, badged capability (similar to the SMC capability)
+that allows threads to ask seL4 to make a RISC-V Supervisor Binary Interface (SBI)
+call on its behalf.
+
+## Motivation
+
+On RISC-V, there is machine mode, supervisor mode, and user mode. Some hardware
+functionality is only available from machine mode and is only accessible through
+SBI.
+
+There exists user-level applications that need to make use of certain M-mode
+only hardware, such as the PMU for doing performance monitoring. One example
+of this is the Foundation's [sel4bench project](https://docs.sel4.systems/projects/sel4bench/).
+Without this RFC, recently released and future RISC-V platforms will not work
+with sel4bench.
+
+## Guide-level explanation
+
+The [SBI](https://github.com/riscv-non-isa/riscv-sbi-doc/tree/master) can be
+invoked via the `ecall` instruction from supervisor mode, in this case seL4.
+
+For each SBI call (excluding the legacy interace which we ignore in this RFC),
+there are two values associated.
+* The Extension ID (EID)
+* The Function ID (FID)
+
+In addition to the EID and FID, there may be up to 5 user-provided arguments.
+Returned are two values, the error indicator and value.
+
+We introduce an 'SBI' capability with an invocation that allows user-space
+to make such SBI calls.
+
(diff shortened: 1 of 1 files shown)


Indanz:
It does say:
> The lower 24 bits of the firmware EID must match the lower 24 bits of the SBI implementation ID
and
> The lower 24 bits of vendor specific EID must match the lower 24 bits of the mvendorid value.

So at least the firmware and vendor EIDs are not arbitrary and we don't have to check them, we can use all the remaining bits for FID.

For the experimental extension it would make more sense to use the remaining bits for EID and not check FID, or to split it 50/50 or something.

kent-mcleod:
> 2. Forget about FID, and only do badging for EID. This works fine in practice, except for a few scary exceptions like Console Read and PMU snapshot shared memory.

Also you can further "attenuate" the rights of a cap by giving it to a user level server that can apply more specific access policy to clients.

Indanz:
> Also you can further "attenuate" the rights of a cap by giving it to a user level server that can apply more specific access policy to clients.

True. We could do EID-only for 32-bit and EID+FID for 64-bit.

lsf37:
It sounds to me like the main objection to the proposed API is "hard to explain to the user". For that: a) I disagree, it is easy to explain with two code examples and b) this is a non-goal -- it doesn't matter if it is hard to explain. What matters is if it achieves security, performance, and verification, and I would argue that it does that better than all of the other variants proposed so far.

I'll go into more details on the variants below, but first: I do agree that the proposal is complex, which does have an influence on verification. We do not have any two-level badging for anything else in the kernel, and while the API is designed carefully to use the existing interfaces that the proofs talk about, it will be necessary to reason through those deeper cap derivation tree levels, which will be annoying and subtle.

The main reason we did not go with just flat badging (AFAIR) was user convenience (also a non-goal, but if everything else is equal, it's still good) and the fact that on rv32, we don't have an existing mechanism to set >32 bit badges on caps even if storage space in the cap is available.

Two new options that sound viable to me that have less complexity are:

5. extend the existing badge mechanism in general with something like a "long" badge, i.e. two data words that are set in the mint/mutate operations, and use that for flat full EID+FID badges on rv32. (Use the normal old badging mechanism for rv64, or also two simultaneous badge words if that is more consistent/convenient -- wouldn't really matter at that point)

6. not support SBI calls on rv32 (unclear to me if that is actually an option -- it was for SMC, but that doesn't mean it is here), and use a flat 64 bit badge on rv64 for EID+FID.

More details on the other variants:

1. adding encoding details for a tighter encoding could work, but it would mean that the kernel now knows a lot more about the structure of SBI than in the current proposal and if we get anything wrong there, we have no guard rails. I.e. we reduce the value of verification with that option. We can still do that if nothing else works, but if we have other options, I don't think it's preferable.

2. absolutely not, see other comment. If we do that, we don't really need to bother with any badging at all, or with verification for that matter.

3. extra syscall: this does not solve anything, and would increase the size of the code change instead of reducing it. It would also be a lot harder to verify. There are two existing mechanisms that we can re-use verification for: normal badging via mint/mutate (which is exactly what that mechanism was designed for) and the control cap/instance cap pattern. In the latter, we get mostly duplication. We did that for SGI caps, because they fit with the existing interrupt handler pattern and could re-use some of that mechanism, but that would not apply here.

4. is a variation of 3 with the same downsides

In more detail, for verification for the new options we have:

5. verification cost on extending the general API to (optionally) long badges would be smaller than reasoning about double-badging as we have in the proposal now. The only interface that changes with long badges in terms of the proofs is argument decoding and the function that sets the cap data. Both are easy. How the cap data is interpreted for the cap derivation tree is independent of all that and if we make it flat, it'd be the same reasoning as we have now, just checking an additional field. That would overall be simpler. It would also open up that mechanism for any future APIs that want more user-modifiable cap data.

6. the simplest of all, of course, but I don't know if that's an option

lsf37:
The TSC has approved variant 5 in the meeting from 3 Dec 2026.

PR #37, opened 2026-04-15
RFC-24: Add support for Intel APICv

dreamliner787-9:
This RFC proposes adding in-kernel support for Intel APICv.

Rendered version here: https://github.com/au-ts/seL4-rfcs/blob/x86_apicv/src/proposed/0240-x86_APICv.md

Prototype kernel implementation: https://github.com/seL4/seL4/compare/master...au-ts:seL4:x86_vmm_dev
Example Microkit integration: https://github.com/seL4/microkit/compare/main...au-ts:microkit:apicv_poc
Example usage in our VMM: https://github.com/au-ts/libvmm/pull/198

### DIFF
--- a/src/proposed/0240-x86_APICv.md
+++ b/src/proposed/0240-x86_APICv.md
@@ -1,0 +1,351 @@
+<!--
+  SPDX-License-Identifier: CC-BY-SA-4.0
+  Copyright 2025 UNSW
+-->
+
+# Intel APICv support for x86 VCPU
+
+<!--
+  To use this template:
+
+ - Fork this repository on GitHub
+ - Choose a number nnn0 greater than any of the nnnx files in
+       ls -l src/*/????-*.md
+
+ - Make a branch <nnn0>-<short-rfc-name>
+ - Copy this template into src/proposed/<nnn0>-<short-rfc-name>.md
+
+ - Replace the title above with the title of your RFC
+ - Choose CC-BY-SA-4.0 above for contributing the RFC (only the template should be MIT)
+ - Add your name/company to Copyright, removing the other ones (they apply to the template)
+
+ - Fill in author, at the end, before you submit, fill in proposal date
+ - Fill in the rest of the sections. It is Ok to leave out sections that do not
+   apply, but don't leave out sections lightly.
+
+ - Make a pull request to <https://github.com/seL4/rfcs> to publish the RFC and
+   start formal discussion.
+
+ - If you are not sure yet how to fill in all sections or want to discuss informally
+   before you start the process, you could post a link to your partially
+   filled in template on your fork to https://sel4.discourse.group/c/rfc-discussion/
+   or start a GitHub gist or other markdown drafting site with the template to
+   work on it incrementally without doing all the setup steps first. Then go
+   through the steps above when you are ready to submit the RFC.
+-->
+
+- Author: Bill Nguyen
+- Proposed: 2026-04-15
+
+## Summary
+
+This RFC proposes adding kernel support that allows userspace to use Intel’s hardware support for Advanced Programmable Interrupt Controller (APIC) virtualisation on x86, commonly referred to as [APICv](https://edc.intel.com/content/www/us/en/design/ipla/software-development-platforms/client/platforms/alder-lake-desktop/12th-generation-intel-core-CPUs-datasheet-volume-1-of-2/010/intel-apic-virtualization-technology-intel-apicv/).
+
+On x86 virtual machines, the guest OS expects to interact with a local APIC for interrupt masking, prioritisation, acknowledgement, and interrupt delivery. This functionality can be emulated entirely in software by a userspace Virtual Machine Monitor (VMM), but this can result in a high rate of VM exits and poor performance for some guests.
+
+Intel APICv is a collection of VT-x features that allow the host CPU to assist with local-APIC virtualisation. Depending on which APICv features are enabled and supported by the host CPU, the CPU can:
+
+- virtualise some guest APIC accesses without involving the VMM,
+- track selected virtual APIC state in hardware, and
+- evaluate and deliver some virtual interrupts without requiring a VM exit.
+
+This RFC does not propose that the kernel implement a virtual APIC. Instead, it proposes that the kernel provide the minimal support needed for userspace VMMs to supply the APICv-related pages and use the relevant Virtual Machine Control Structure (VMCS) controls safely through the existing VCPU interface.
+
+The initial scope of this RFC covers these APICv-related features:
+
+- Use TPR shadow
+- Virtualise APIC accesses
+- APIC-register virtualisation
+- Virtual-interrupt delivery
+- Virtualised End-of-Interrupt (EOI)
+
+Other related features, such as x2APIC virtualisation, virtual interrupt posting, and IPI virtualisation, are left for future work.
+
+This RFC does not break backward compatibility.
+
+## Motivation
+
+- Why are we doing this?
+
+An interrupt controller is a mandatory requirement for any commodity Operating Systems that you would run inside a Virtual Machine. While virtualising the APIC in pure software is sufficient for Linux guests, we have found that virtualising Windows without APICv resulted such poor performance that the guest was unusable.
+
+One of the specific issues we ran into was related to the Task Priority Register (TPR), also known as Control Register 8 (CR8). The TPR is used by the CPU to determine which classes of interrupts are currently allowed to be delivered. In other words, it acts as a simple interrupt priority threshold: interrupts at or below the threshold are masked, while interrupts above it may be delivered by the local APIC.
+
+We measured that the Windows NT kernel updates the TPR very frequently, approximately 50,000 times per second on our setup. In a software APIC model, each TPR access can require a VM exit so that the VMM can emulate the operation and update the virtual APIC state. This corresponds to approximately 50,000 VM exits per second, or roughly 200,000 context switches per second for just emulating the TPR (guest -> seL4 -> VMM -> seL4 -> guest)!
+
+We have not seen Linux use the TPR anywhere near as much as Windows, which is why the performance on Linux guests has been acceptable.
+
+When we switched to using Intel APICv, those TPR updates could be handled by the CPU without requiring a VM exit to the VMM. This eliminated a major source of VM exits. On our VMM, Windows 10 boot time dropped from roughly 15-20 minutes to approximately 1-2 minutes, and the resulting desktop became usable.
+
+TPR virtualisation is only part of the motivation. APICv also reduces exits on the interrupt delivery path more generally.
+
+With a purely software APIC, the typical flow for delivering a virtual interrupt looks roughly like this:
+
+1. An external interrupt is raised.
+2. If the guest VCPU is currently running, the interrupt causes a VM exit so the kernel can schedule the VMM.
+3. The VMM receives the notification and examines the virtual APIC state to determine whether the guest is currently able to take the interrupt.
+4. If the guest has interrupts disabled, or if the interrupt priority rules do not currently permit delivery, the VMM must defer injection and resume the guest.
+5. At some later point in time, another VM exit occurs when the VMM gets another opportunity to attempt delivery when the architectural rules are satisfied.
+6. Once the interrupt is finally injected, the guest eventually signals completion by writing to the APIC End-Of-Interrupt (EOI) register, which itself triggers another VM exit so that the VMM can update interrupt-controller state.
+
+This means that a single virtual interrupt involves several VM exits and multiple round-trips between guest, kernel, and VMM.
+
+With APICv, most of this work can be offloaded to the CPU. The VMM can mark a virtual interrupt as pending via and resume the guest, while the CPU tracks the relevant virtual APIC state and performs interrupt evaluation and delivery directly in VMX non-root mode where possible. This reduces VM exits and lowers the overhead for interrupt-heavy guests.
+
+- What problem does it solve? What use cases does it support?
+
+This RFC allows more performant guests on x86, in particular, it allows for Windows guests to have acceptable performance.
+
+It is particularly useful for guests that update the virtual APIC state frequently or receive a large number of interrupts.
+
+- What is the expected outcome?
+
+The expected outcome is that userspace VMMs on seL4 can use Intel APICv support to offload parts of local-APIC virtualisation to the CPU, substantially reducing VM exits and improving guest performance without moving APIC policy or virtualisation logic into the kernel.
+
+## Guide-level explanation
+
+<!-- Explain the change or feature as you would to another user of seL4. -->
+<!-- This section should:
+
+- clearly outline new named **concepts**
+- provide some **examples** of how it will be used, and
+- explain how **users** should think about it.
+
+You should describe the change in a manner that is clear to both existing seL4
+users and to new users of the ecosystem. For instance, if the change is to the
+seL4 API, this section should include the prose of the corresponding new
+paragraphs or sections int the seL4 manual.
+
+Any changes that existing users need to make either as a result of the RFC or
+such that the can use the feature should be clearly stated here. -->
+
+<!-- Document referenced (will be referenced in this RFC as "the Intel manual"):
+Intel® 64 and IA-32 Architectures Software Developer’s Manual
+Combined Volumes: 1, 2A, 2B, 2C, 2D, 3A, 3B, 3C, 3D, and 4
+Order Number: 325462-088US
+June 2025 -->
+
+Support for this feature is gated behind a new configuration flag:
+
+`INTEL_APICV`
+
+By default, this option is OFF.
+
+If the option is OFF, there is no change to kernel behaviour.
+
+If the option is ON, the kernel is built with support for APICv-related checking and exposes a new system call for configuring the APICv pages. The userspace VMM may then use whichever supported APICv controls are available on the current host CPU. If the host CPU does not support all of the APICv features implemented in this RFC, the kernel will refuses to boot.
+
+To use APICv, userspace must provide two 4 KiB frames to the kernel using a new system call:
+
+`seL4_X86_VCPU_SetAPICvPages()`
+
+These frames are used as:
+
+- the APIC-access page, and
+- the virtual-APIC page.
+
+Conceptually:
+
+The APIC-access page is the page through which xAPIC-style memory-mapped APIC register accesses are virtualised.
+The virtual-APIC page is a page used by VT-x to store selected virtual local-APIC state.
+
+The APIC-access page should be mapped into guest RAM through EPT at the location expected for guest APIC accesses. Conventionally `0xFEE0_0000`.
+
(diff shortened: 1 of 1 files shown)


midnightveil:
Comment on the implementation: you also need to store the `deriveCap` from the `decode` somewhere in a CNode (like with the various TCB methods that do `cteInsert`. is there an equivalent mini-cnode in the VCPU?). Otherwise, you've still created a use-after-free like ability. Just the `deriveCap` isn't good enough as it does not store permanent state.

Do we have a place to store those caps at this time? Does that need to be added?

Indanz:
Why is a new syscall needed? Why can't you just add the `VMX_CONTROL_APIC_ACCESS_ADDRESS` and `VMX_CONTROL_VIRTUAL_APIC_ADDRESS` to the set of registers that are context switched and modifiable by the VMM?

`invokeVCPUSetAPICvPages` just sets those two registers and `decodeVCPUSetAPICvPages` doesn't do any interesting checks anyway.

Then the diff is really minimal and there isn't really a need for an RFC at all.

Edit: Ah, because those registers take a host physical address, not a guest virtualised address. How inconvenient.

Well, in that case, you totally forgot about cleanup, like Julia already pointed out.

Indanz:
You don't need to derive caps and embed them anywhere. Such caps are currently not used to cleanup the TCB when the other objects goes away. 

In this case, if the TCB goes away, you don't have to do anything. But you do need to clear the VMX registers when the page objects goes away. We can't look up the TCB/VCPU by ASID, as the VSpace may be shared by multiple instances.

In many ways it's similar to the IPC buffer. There the problem is solved by checking the IPC buffer cap every time it's used, but that approach doesn't work here.

If the cap is embedded, on delete of a frame cap you could check if the cte_t is embedded (special flag in cte_t or cap). From there you could retrieve the encompassing TCB object. (This trick could also be used to avoid checking the IPC buffer cap on each call.) However, `finaliseCap` doesn't get the cte_t, only the cap data, it's necessary to change `finaliseSlot` to handle this case. I implemented this approach [here](https://github.com/seL4/seL4/compare/master...Indanz:seL4:maput) for an unfinished proof-of-concept.

If the frame is part of the VCPU you still need a way to get derived frame caps. Such derived frame caps must be child objects of the VCPU, otherwise you get use-after-free when deleting the VCPU object. But that's the only tricky bit of this approach. All-in-all this approach seems much simpler, as you don't need to modify common code.

Indanz:
I think https://github.com/seL4/seL4/pull/1098 is more important, as that fixes security problems, while this is just an optimisation.

Ivan-Velickovic:
> I think [seL4/seL4#1098](https://github.com/seL4/seL4/pull/1098) is more important, as that fixes security problems, while this is just an optimisation.

While we want to have that PRs changes as well, isn't it completely separate to the RFC?

lsf37:
I'm unclear on what exactly needs to be cleaned up when. There are two scenarios to consider: a) the frame caps that the user has provided are deleted and b) the VPCU object is deleted. I don't think TCBs are involved, although maybe that is different on x86 (I'm mostly familiar with the Arm model). For a), how do we find the connection to the VCPU where we changed hardware state that causes writes/reads to those frames? For b) how do we find the corresponding frames? (do we need to? currently I don't think so, but as soon as a) is fixed there is likely information that needs to be cleaned up from the other direction).

The currently implementation in https://github.com/seL4/seL4/compare/master...au-ts:seL4:x86_vmm_dev is clearly broken for the case a), though. At the very least integrity is broken, but I'm pretty sure I can make the kernel crash with it. I think this is what Julia meant, but just storing derived caps does not fix anything yet. We also don't really need to store caps for tracking connections, a pointer is good enough, but it needs to be pointers in both directions.

For Indan's proposal: we definitely do not want flags in CTEs. Maybe there is something we can do with caps analogous to the mapping information for frames in page tables, but it does sound like a hack.

In seL4 in general, if two objects are bound to each other like here, there needs to be storage available for that binding on each side. If one of those objects is just a frame, that is not possible, and so I don't see how this proposal can work. The user is simply not providing enough storage for the kernel's book keeping.

Indanz:
> While we want to have that PRs changes as well, isn't it completely separate to the RFC?

Technically yes, but from a manpower point of view, no, as both are x86 virtualisation related.

> In seL4 in general, if two objects are bound to each other like here, there needs to be storage available for that binding on each side. If one of those objects is just a frame, that is not possible, and so I don't see how this proposal can work. The user is simply not providing enough storage for the kernel's book keeping.

The missing information can be provided implicitly by embedding the `cte_t` in the object you are looking for.

> For Indan's proposal: we definitely do not want flags in CTEs.

Alternatively, a pointer to the slot could be passed to `finaliseCap`, but my impression was that that would complicate existing already complicated proofs, while adding an extra cleanup step seemed simpler.

Once you do the above trick, it's much cleaner to mark embedded CTEs explicitly with a flag in `mdb_node_t` than to infer that from the cap data, because that keeps the generic cap code generic. We already have `mdbRevocable` and `mdbFirstBadged`, this would be a new `mdbEmbedded`.

The flag would be used in `finaliseSlot` to check whether a cap might need additional cleanup before calling `finaliseCap`. It would replace checks like [this](https://github.com/Indanz/seL4/blob/115d1347b3ea93598ea15a8ba1e04a3a3c363738/src/object/cnode.c#L624) with a generic check and then call `unlink_slot` or something like that, which does the cap type check and clears the `mdbEmbedded` flag.

> Maybe there is something we can do with caps analogous to the mapping information for frames in page tables, but it does sound like a hack.

The only clean alternative I can think of is to expand MDBs to have two next and prev pointers so multi-parent relationships can be tracked properly. Ideally this could replace the current ASID hacks we have too. But it would double the size of caps, which is not entirely bad either, because space in caps is already tight.

lsf37:
> > In seL4 in general, if two objects are bound to each other like here, there needs to be storage available for that binding on each side. If one of those objects is just a frame, that is not possible, and so I don't see how this proposal can work. The user is simply not providing enough storage for the kernel's book keeping.
> 
> The missing information can be provided implicitly by embedding the `cte_t` in the object you are looking for.

No, that is not a good way to embed that information. Embedding ctes in random objects can violate all kinds of assumptions that other code makes. Since this is not verified code I would do something like this only under extreme circumstances. The proofs make essential use of the fact that CTEs can only live in CNodes and TCBs (and that was already not a good idea if you ask me), and extending that breaks every argument that makes that case distinction. That's hundreds. That means, if this feature is without proof (and it has to be, because VT-x is already without proof), there is no reason to believe that all those hundreds of arguments would extend naturally. That in turn means this feature would be messing with the central authority mechanism of the kernel and could break not only the thing it extends, but everything else.
 
> > For Indan's proposal: we definitely do not want flags in CTEs.
> 
> Alternatively, a pointer to the slot could be passed to `finaliseCap`, but my impression was that that would complicate existing already complicated proofs, while adding an extra cleanup step seemed simpler.
> 
> Once you do the above trick, it's much cleaner to mark embedded CTEs explicitly with a flag in `mdb_node_t` than to infer that from the cap data, because that keeps the generic cap code generic. We already have `mdbRevocable` and `mdbFirstBadged`, this would be a new `mdbEmbedded`.

Sorry, but all of that sounds absolutely horrible and is not in line with the existing design.

> The flag would be used in `finaliseSlot` to check whether a cap might need additional cleanup before calling `finaliseCap`. It would replace checks like [this](https://github.com/Indanz/seL4/blob/115d1347b3ea93598ea15a8ba1e04a3a3c363738/src/object/cnode.c#L624) with a generic check and then call `unlink_slot` or something like that, which does the cap type check and clears the `mdbEmbedded` flag.

That kind of check you are referencing is against the CTE design in seL4 and should be eliminated, not generalised. CTE handling should be position independent and not rely on which object the CTE is in. It creates yet another special case for CTEs and yet another case where normal CTE operations do not apply (no copy/move/etc). It makes the API less uniform and even harder reason about. All of that is in the wrong direction.

> > Maybe there is something we can do with caps analogous to the mapping information for frames in page tables, but it does sound like a hack.
> 
> The only clean alternative I can think of is to expand MDBs to have two next and prev pointers so multi-parent relationships can be tracked properly. Ideally this could replace the current ASID hacks we have too. But it would double the size of caps, which is not entirely bad either, because space in caps is already tight.

Yes, that's one of the reasons caps are just the wrong vehicle for tracking such information. There is no reason to believe that shoehorning other object dependency information into the authority mechanism would make the design better.

The instances where the dependency tracking works fine is where objects have storage for explicitly tracking it. That is what the design should have. Pure frames without extra storage are fundamentally problematic for that.

Frankly, having them embedded in the VCPU would be drastically simpler and I don't buy the performance waste argument. Either you want to use this feature or you don't. Why would you want to use it only for some VCPUs in the same system? But even that could be achieved: you could have two types of VCPU sizes in the same system and you create the one you want at runtime.  Sounds like 0 cost to me without creating yet another case of single-use object binding and complexity where none is needed.

People seem to think for some reason that object binding is a good idea, but every single case of it has ended in a complexity disaster for what it achieves. Implementing it with caps would make it even worse. If we can avoid it, then we definitely should.

lsf37:
> Frankly, having them embedded in the VCPU would be drastically simpler and I don't buy the performance waste argument. Either you want to use this feature or you don't. Why would you want to use it only for some VCPUs in the same system? But even that could be achieved: you could have two types of VCPU sizes in the same system and you create the one you want at runtime. Sounds like 0 cost to me without creating yet another case of single-use object binding and complexity where none is needed.

The real argument against this design variant is that the user can no longer map those frames. Can that be solved some other way? Is access to those frames performance critical or could it be via the kernel?

dreamliner787-9:
> Frankly, having them embedded in the VCPU would be drastically simpler

Yes I agree, it was under alternative solution 1 of the RFC as well to solve the resources tracking problem unfolding above. But I didn't implement that solution for the main proposal due to the performance problem below.

> Is access to those frames performance critical or could it be via the kernel?

Unfortunately the VMM still need to access the virtual APIC page to emulate certain functions of the APIC that were not hardware accelerated (e.g. the APIC timer, which is used by Linux for scheduling).

You can see the an example of how often and when a VMM need to access the virtual APIC page in the [virtual APIC code of our VMM](https://github.com/au-ts/libvmm/blob/x86_64_upstream_simple/src/arch/x86_64/apic.c).

For example, when a guest kernel sets a timeout in the APIC timer by writing the initial count to `REG_LAPIC_INIT_CNT`. We need to read that value from the virtual APIC page plus `REG_LAPIC_DCR` and `REG_LAPIC_TIMER` to ascertain now long in host time the timer goes for.

Similarily when a guest kernel reads the current count we also need to read multiple APIC registers to work out what the counter should be and return it to the guest.

If we do this virtual APIC register reads and writes via an invocation it would incur significant performance penalty for the guest. Unless upon a VM Exit due to APIC write, we deliver the relevant virtual APIC registers value via message registers to the VMM to reduce the amount of context switches. Or move some of this logic into the kernel which is bad.

lsf37:
> Unfortunately the VMM still need to access the virtual APIC page to emulate certain functions of the APIC that were not hardware accelerated (e.g. the APIC timer, which is used by Linux for scheduling).

Ok, so we should investigate if we can make the frames available to the user.

I haven't thought this through yet, but what about allowing the use of VCPU caps in map/unmap operations? We might have to add the same kind of mapping info to VCPU caps that frame caps have now so that they can be unmapped again (not exactly nice, but there would be space in the cap). When the VCPU object is deleted, the frame would have to be unmapped the same way other frames are unmapped on deletion. There is no separate frame object, so we never have the case where the frame is deleted and we need to find the VCPU object or the VCPU is deleted and we need to fine the frame.

We can have the case we also get for frames in the other design that the page table is deleted in which the frame mapping resides. This would mean the VPCU page is no longer accessible to the user, but the hardware would still be writing/reading to the physical frames. I think that would be fine -- it's the same for as in other shared frame situations and no different in the other design. 

The wrinkle is that you need to handle two frames instead of one, but that seems relatively straightforward to me compared to the other design where we would need to change general cap mechanisms. This one is only adding code for things that are new.

It doesn't sound entirely appealing to me, but better than separate objects with the problems of separate lifetimes.

Indanz:
> I haven't thought this through yet, but what about allowing the use of VCPU caps in map/unmap operations? We might have to add the same kind of mapping info to VCPU caps that frame caps have now so that they can be unmapped again (not exactly nice, but there would be space in the cap).

For this the ASID and the virtual address the page is mapped at need to be stored. The x86 VCPU caps have plenty of space left for that.
 
> The wrinkle is that you need to handle two frames instead of one, but that seems relatively straightforward to me compared to the other design where we would need to change general cap mechanisms. This one is only adding code for things that are new.

Selecting between the two pages should be trivial.

I thought directly mapping wouldn't work because you need to be able to map the same VCPU page in multiple VSpaces (both in the guest VSpace and the VMM), but I forgot that you can just copy the VCPU cap. Hence my proposal above to give a way to derive page caps from a VCPU cap instead.

dreamliner787-9:
> but what about allowing the use of VCPU caps in map/unmap operations?

> For this the ASID and the virtual address the page is mapped at need to be stored.

Just confirming before I make a new implementation. So we would embed the 2 extra pages in the VCPU object like so:
```diff
struct vcpu {
    /* Storage for VMCS region. First field of vcpu_t so they share address.
     * Will use at most 4KiB of memory. Statically reserve 4KiB for convenience. */
    char vmcs[VCPU_VMCS_SIZE];

+#ifdef CONFIG_INTEL_APICV
+    /* Storage for the 2 pages needed for Intel APICv. Place it after the VMCS
+     * to maintain 4KiB alignment. */
+    char apicv_access[VCPU_APICV_ACCESS_PAGE_SIZE];
+    char apicv_virtual_apic[VCPU_APICV_VIRTUAL_APIC_PAGE_SIZE];
+#endif

    word_t io[VCPU_IOBITMAP_SIZE / sizeof(word_t)];

    /* Place the fpu state here so that it is aligned */
    user_fpu_state_t fpuState;

+#ifdef CONFIG_INTEL_APICV
+    asid_t apicv_access_mapped_asid;
+    asid_t apicv_virtual_apic_mapped_asid;
+    word_t apicv_access_mapped_gpa;
+    word_t apicv_virtual_apic_mapped_vaddr;
+#endif

```

Then we add 4 new x86 VCPU invocations:
- `seL4_X86_VCPU_MapApicAccessPage(seL4_X86_VCPU _service, seL4_X86_EPTPML4 eptpml4, seL4_Word gpa)`
- `seL4_X86_VCPU_MapVirtualApicPage(seL4_X86_VCPU _service, seL4_CPtr vspace, seL4_Word vaddr, seL4_CapRights_t rights, seL4_X86_VMAttributes attr)`
- `seL4_X86_VCPU_UnmapApicAccessPage(seL4_X86_VCPU _service)`
- `seL4_X86_VCPU_UnmapVirtualApicPage(seL4_X86_VCPU _service)`

These will have similar semantics to the normal map/unmap operations. Have I understood you both correctly? @lsf37 @Indanz

Indanz:
> +    asid_t apicv_access_mapped_asid;
> +    asid_t apicv_virtual_apic_mapped_asid;
> +    word_t apicv_access_mapped_gpa;
> +    word_t apicv_virtual_apic_mapped_vaddr;

If each page only needs to be mapped once, then yes, having this information in the VCPU object itself would work and would be the better solution.

However, once it must be possible to map the same page multiple times, then you need to store the ASID + virtual address in the VCPU cap itself instead of in the VCPU object.

>     * `seL4_X86_VCPU_MapApicAccessPage(seL4_X86_VCPU _service, seL4_X86_EPTPML4 eptpml4, seL4_Word gpa)`
> 
>     * `seL4_X86_VCPU_MapVirtualApicPage(seL4_X86_VCPU _service, seL4_CPtr vspace, seL4_Word vaddr, seL4_CapRights_t rights, seL4_X86_VMAttributes attr)`

Personally I would not add new syscalls, but use the existing map ones and extend them to also work on VCPU caps. And then add two special flags to `seL4_X86_VMAttributes`, one for APIC access page and one for the virtual APIC page.

dreamliner787-9:
> Personally I would not add new syscalls, but use the existing map ones and extend them to also work on VCPU caps. And then add two special flags to seL4_X86_VMAttributes, one for APIC access page and one for the virtual APIC page.

Ah cool I didn't think of that, hence I proposed adding the new syscalls. But the flag approach is a lot better.

midnightveil:
> Personally I would not add new syscalls, but use the existing map ones and extend them to also work on VCPU caps. And then add two special flags to seL4_X86_VMAttributes, one for APIC access page and one for the virtual APIC page.

I don't think this is possible — for (bad and should be changed) reasons page mapping operations are specified as invocations on the *frame* cap, not the page table. So we can't add to the existing frame cap invocations as we won't have a frame capability to do an invocation on.

Though it might make more sense to add these as invocations on the page table capabilities if we're adding new syscalls anyway, though that would make the API quite confusing.

Indanz:
> I don't think this is possible — for (bad and should be changed) reasons page mapping operations are specified as invocations on the _frame_ cap, not the page table. So we can't add to the existing frame cap invocations as we won't have a frame capability to do an invocation on.

Yes, but that doesn't matter in this case. It just means that the VCPU object has to implement the `X86PageMap`, `X86PageUnmap` and `X86PageMapEPT` invocations. As those identifiers are unique, any object can implement them.

Edit: To clarify, the API xml needs no change, other than perhaps documentation. You don't add these invocations to the VCPU object in the XML, as that indeed won't work.

Whether the VCPU object does its own thing or redirects it to the page code (with or without a temp frame cte_t) is an implementation detail.

lsf37:
> These will have similar semantics to the normal map/unmap operations. Have I understood you both correctly? @lsf37 @Indanz

Sorry, have been travelling for a while. Yes, embedding the pages like this is what I meant. As Indan already indicated, the mapping info should probably go into the caps (like we have it now in the frame caps), and I do think you are right that we will unfortunately need these as invocations on the VCPU caps because as Julia said, map/unmap is an invocation of the frame cap not the page table cap.

It might still be possible to share the code between them in the way I think Indan means -- not sure if that ends up more or less messy than the duplication we would get otherwise. We'll probably have to try it and see.

(Edit: the main trick in sharing the code between VCPU and frame caps will be to do this in a way that doesn't break verification on non-VT-x configs)

Indanz:
> It might still be possible to share the code between them in the way I think Indan means -- not sure if that ends up more or less messy than the duplication we would get otherwise. We'll probably have to try it and see.

Yes, hard to tell beforehand if my suggestion is better or worse than just duplicating code.

My idea was to call the page decode functions directly from the VCPU code. But to do that, you need a temporary local cte_t with a page cap setup properly and pass that instead of the vcpu cap/cte_t. Then it's just a wrapper that does some VCPU related bookkeeping around the page calls.

However, looking at the actual code, for invoke it seems much simpler to call `createSafeMappingEntries_PTE` and `invalidatePageStructureCacheASID` directly. Decode has a lot of boiler plate that would need to be duplicated though (around 100 lines of code).

Maybe the useful checking code in `decodeX86FrameInvocation` can be somehow shared, but not without verification changes. All in all unclear which approach is better.

PR #41, opened 2026-08-06
RFC-26: Microkit: lifting the 62-channel limit

midnightveil:
Please see RFC description.

[View the rendered version here](https://github.com/au-ts/seL4-rfcs/blob/julia/microkit-api-changes/src/proposed/0260-microkit-api-changes.md).

Related: https://github.com/seL4/microkit/issues/526, https://github.com/seL4/microkit/issues/510, https://github.com/seL4/microkit/issues/362.

### DIFF
--- a/src/proposed/0260-microkit-api-changes.md
+++ b/src/proposed/0260-microkit-api-changes.md
@@ -6,24 +6,27 @@
   Based on the Rust RFC template at <https://github.com/rust-lang/rfcs>
 -->
 
-# Microkit API/Entrypoint Changes
+# Microkit: lifting the 62-channel limit
 
 - Author: Julia Vassiliki, UNSW
 - Proposed: 2026-08-06
 
 ## Summary
 
-We propose decoupling the 1:1 mapping the channel IDs used for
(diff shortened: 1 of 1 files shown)


src/proposed/0260-microkit-api-changes.md
```diff
@@ -0,0 +1,744 @@
...
+<!-- We replace <channel> elements with a new
+     <notification_channel> element to indicate
+     that these apply only to notifications (and IRQs)
+     In theory, this is unnecessary, but we'd need to
+     add something like ppcall_id="x". Maybe better? -->
+<notification_channel>
+    <!-- virtualiser would call microkit_notify(1)
+         to talk to PD "a".
+
+         - recv_id must lie in the range [0, 62],
```
Courtney3141, line 146: Did you switch rev_id and send_id here?

midnightveil: I don't think so? Receive ID corresponds to the ID in the badge that gets set during `seL4_Recv`.

midnightveil:
I've removed the details that now exist in RFC-27. This is probably worth a run over for coherency, which I'll do later once I can effectively do it.

PR #42, opened 2026-08-14
RFC-27: Microkit Entrypoint/Deferred API Changes

midnightveil:
Please see RFC description.

This has been split off from #41 in order to focus these on narrower aspects. There is some relevant discussion there.

[View the rendered version here](https://github.com/au-ts/seL4-rfcs/blob/rfcs-notified-deferred/src/proposed/0270-microkit-entrypoint-deferred-changes.md).

### DIFF
--- a/src/proposed/0270-microkit-entrypoint-deferred-changes.md
+++ b/src/proposed/0270-microkit-entrypoint-deferred-changes.md
@@ -1,0 +1,352 @@
+<!--
+  SPDX-License-Identifier: CC-BY-SA-4.0
+  Copyright 2024 seL4 Project a Series of LF Projects, LLC.
+  Copyright Rust Language Community
+
+  Based on the Rust RFC template at <https://github.com/rust-lang/rfcs>
+-->
+
+# Microkit Entrypoint & Deferred API Changes
+
+- Author: Julia Vassiliki, UNSW
+- Proposed: 2026-08-14
+
+## Summary
+
+We propose changing the `void notified(channel) { }` entrypoint (called once
+per-channel from the notification badge) into `notified(channel_set) { }` (called
+once with the notification badge).
+
+We also propose removing the `microkit_deferred_signal` APIs, and utilising the
+return value `microkit_notified_ret_t` of the new `notified(channel_set)`
+entrypoint to implemenet syscall-combining.
+
+## Motivation
+
+This is motivated in part by the GitHub issue [Deferred notify and IRQ ack may
+be overwritten](https://github.com/seL4/microkit/issues/510).
+
+The `microkit_deferred_notify()` API allow users to tell the Microkit handler
+loop to utilise the combined `seL4_NBSendRecv` syscall to combine two syscalls
+into one; in contrast to the `microkit_notify()` function which immedately performs
+`seL4_Signal`. There is also a `microkit_deferred_irq_ack()`. Until recently,
+calling either of these functions twice would overwrite the same global
+`microkit_deferred_send` variable storing an `seL4_MessageInfo_t` and
+silently discard your previous deferred syscall invocations, now, they perform
+the previous deferred call. This is considered a user error, but it's not very
+obvious unless one knows internals of how the Microkit handler loop works.
+Additionally, the `deferred` APIs can be called from the `fault()` or `protected()`
+entrypoints, but they don't make sense here: these endpoints have replies already,
+so it would be trying to combine two sends and a recv into a SendRecv call.
+
+Something surprising is then that it's one `deferred_ack` call *per set of `notified()`*,
+because one exit from `seL4_Recv` can call `notified()` multiple times; but you have
+no way of knowing whether or not it is the "last" notified. This is both
+confusing as a user (to use it correctly) but also makes verification harder.
+
+The second part is that there is an implicit ordering of the notification handling;
+currently it handles the lowest channel ID first, and goes down. The fact that
+you may have multiple pending notifications is also not exposed to you, as the
+user. This motivates the `notified(ch) -> notified(channel_set)` change by itself,
+which then allows the new deferred API.
+
+## Guide-level explanation
+
+### Previous (current) API
+
+```c
+/* PD "virtualiser" - user code */
+
+#define CH_DRIVER (0)
+#define CH_PD_A (1)
+#define CH_PD_B (2)
+
+void notified(microkit_channel ch) {
+    if (ch == CH_DRIVER) {
+        /* it doesn't really make sense to do deferred_notify here */        */
+        do_stuff();
+        microkit_notify(CH_PD_A);
+        /* if we were to always do a deferred, if the ordering
+           between notify(1) and notify(2) matters the deferred
+           API has non-obvious behaviour */
+        microkit_notify(CH_PD_B);
+    } else if (ch == CH_PD_A || ch == CH_PD_B) {
+        do_other_stuff();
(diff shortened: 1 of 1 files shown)


midnightveil:
For time-context, please see the comments prior to [this one by me](https://github.com/seL4/rfcs/pull/41#issuecomment-5210368993) for other discussions. I will edit this comment later to include copies of the relevant sections.

src/proposed/0270-microkit-entrypoint-deferred-changes.md
```diff
@@ -0,0 +1,352 @@
...
+        microkit_deferred_notify(CH_DRIVER);
+    }
+}
+```
+
+### Proposed (new) API
+
+The `notified()` API takes an `seL4_Word` sized argument "channel_set" that
+is a one-hot encoding of the channels; i.e. the literal badge (with one bit
+masked off) used for the notification. It then also returns a tagged enum
```
zaklogician, line 84: I suspect that, if we have the `notified(badge)` API, the `NBSendRecv` mechanism becomes superfluous, and can just be removed instead of creating a new return-type gadget for invoking it.

src/proposed/0270-microkit-entrypoint-deferred-changes.md
```diff
@@ -0,0 +1,352 @@
...
+        microkit_notify(CH_PD_B);
+    }
+
+    /* Note the "if" here, not elif */
+    if (channel_set & (BIT(CH_PD_A) | BIT(CH_PD_B))) {
+        do_other_stuff();
+        /* on return from notified(), we will do this signal */
+        return microkit_notified_ret_signal(CH_DRIVER);
+        /* a driver could use this variant */
+        return microkit_notified_ret_irq_ack(...);
```
zaklogician, line 118: With this API, after this return, the next iteration of the handler loop begins, and a conditional is used to invoke `NBSendRecv` instead of `Recv`.

If instead of the proposed return types we just had `void notified(seL4_Word badge)`, we could get nigh-identical behavior using
```
  microkit_irq_ack(...);
  return;
```

This means that there is no need to introduce these returns types: one can always just rewrite a deferred notify as a "notify before return". At least unless there is a significant and necessary performance advantage to `NBSendRecv` instead of two calls. One has to benchmark, but I doubt it matters.

If the user wants a specific notify-based-on-returned-value mechanism for some psychological reason, they can always implement it themselves.

zaklogician: Note: I do understand that if a PD L calls `microkit_notify(..)` on a channel that maps to a higher-priority PD H, this may wake H up and cause the kernel to immediately preempt L in order to execute H.

I guess this could lead to measurable performance impact in a situation where H in turn immediately notifies L. But
1. I doubt it
2. the return-type-based solution can also lead to worse performance in various cases, in particular it might introduce an extra branch.

midnightveil: The deferred API - and the return API - is all to make use of the ability to do a signal + wait together via NBSendRecv.

I also don't know if it matters, but Gernot definitely wants to use the feature. It would in theory saves cycles, but whether it makes a difference overall, IDK.

The cost of a branch is going to be massively less than the cost of an extra context switch.

zaklogician: > The cost of a branch is going to be massively less than the cost of an extra context switch.

That's clear, but its not a one-to-one comparison: the branch cost is paid by everybody using Microkit (in parcular all PDs on a system), whereas the extra context switch is fairly niche.

The original reason for introducing deferred wasn't pure performance: it is simply the only way to do certain common things with the current loopy notified api.

midnightveil: Offline discussion: proposal from Zoltan is to remove the `have_signal` branch entirely, the return value can avoid that branch.

> The original reason for introducing deferred wasn't pure performance: it is simply the only way to do certain common things with the current loopy notified api.

Right. You previously had to use it if you wanted to only signal once across all the notifies because you didn't know how many times you would be called, leading to over-signalling otherwise.

This was not my understanding of why this was added: see the description in this PR which explicitly states it is for performance: https://github.com/seL4/microkit/pull/183.

zaklogician: Nice, that gives perf as the _sole_ justification. Thanks for looking into this. I might do some archeology to try and figure out what the first use cases were, and how they actually affected perf.

src/proposed/0270-microkit-entrypoint-deferred-changes.md
```diff
@@ -0,0 +1,352 @@
...
+  particularly imaginative, description.
+
+* For the `fault()` entrypoint, we could replace the existing `bool` return and
+  `seL4_MessageInfo_t *tag` argument with the same mechanism as `notified`:
+  this would allow `fault()`, in the case one is not replying to a fault, to do a
+  combining of the calls. It would also be more consistent.
+
+  i.e. `tagged_enum microkit_fault_ret { reply, ntfy(cap), irq_ack(cap), none }`.
+
+* Calling it `badge` instead of `channel_set`?
```
dreamliner787-9, line 352: I prefer `channel_set`.

src/proposed/0270-microkit-entrypoint-deferred-changes.md
```diff
@@ -0,0 +1,352 @@
...
+  maybe a `notified_single()` variation to make the fix a simple rename, or
+  something more complex (two variations of `libmicrokit`?) if we really want
+  to keep backwards compatibility.
+  We can also retain the deferred APIs and make them just aliases of the
+  standard `microkit_signal` API.
+
+* The extra logic with checking bitmasks manually in `notified()` might
+  be more complicated; previously you did not have to do bitmasks.
+  We could resolve this with helper functions, like in the `rust-seL4`
+  `ChannelSet` API, although C makes that somewhat more difficult.
```
dreamliner787-9, line 261: Can't we just provide something like:
```c
typedef uint64_t microkit_channel_set_t;
bool microkit_channelset_is_set(microkit_channel_set_t channel_set, uint64_t channel) {
    return channel_set & (1ul << channel);
}
```

This way, it is not too bad for Microkit users to adapt the system to the new API. Unless I'm over looking something.

midnightveil: Yes we can.

src/proposed/0270-microkit-entrypoint-deferred-changes.md
midnightveil: See sibling discussion: https://github.com/seL4/rfcs/pull/42#discussion_r3793320015

I'm going to close this, just to keep the discussion in the sibling thread.

src/proposed/0270-microkit-entrypoint-deferred-changes.md
Indanz: There is no fastpath for `NBSendRecv`, compared to signal (though not verified), so gain is less than you'd expect.

Indanz: The main problem I have with the deferred API is that it either should be done automatically, so users don't need to be aware of it and all code benefits from the optimisation, or not at all.

If you add a Microkit context parameter to all Microkit calls, then you won't need globals and could do deferral transparently.

